PR #55, opened 2019-03-27
Arbitrary hash docs

nhemsley:
Documentation update to explain how to use arbitrary hashes in a json attribute.

There are no tests to cover this functonality, and it is probably dangerous.

Any thoughts?


nhemsley:
Oops, broken branch.

PR #206, opened 2023-06-19
Support store keys in polymorphic models

rdubya:
This addresses issue https://github.com/jrochkind/attr_json/issues/205

The changes look more extensive than they are, I basically moved the logic from the cast into a shared method that can selectively call either `cast` or `deserialize` on the type to rehydrate the model. Open to suggestions if you prefer different naming or don't like the `public_send` approach.

### DIFF
--- a/lib/attr_json/type/polymorphic_model.rb
+++ b/lib/attr_json/type/polymorphic_model.rb
@@ -143,6 +143,12 @@ module AttrJson
 
       protected
 
+      # We need to make sure to call the correct operation on
+      # the model type, so that we get the same result as if
+      # we had called the type directly
+      #
+      # @param v [Object, nil] the value to cast or deserialize
+      # @param operation [Symbol] :cast or :deserialize
       def cast_or_deserialize(v, operation)
         if v.nil?
           v
@@ -157,6 +163,8 @@ module AttrJson
         end
       end
 
+      # @param hash [Hash] the value to cast or deserialize
+      # @param operation [Symbol] :cast or :deserialize
       def model_from_hash(hash, operation)
         new_hash = hash.stringify_keys
         model_name = new_hash.delete(type_key.to_s)
@@ -167,7 +175,13 @@ module AttrJson
 
         raise_bad_model_name(model_name, hash) if type.nil?
 
-        type.public_send(operation, new_hash)
+        if operation == :deserialize
+          type.deserialize(new_hash)
+        elsif operation == :cast
+          type.cast(new_hash)
+        else
+          raise ArgumentError, "Unknown operation #{operation}"
+        end
       end
 
       def raise_missing_type_key(value)


jrochkind:
Hey, thanks for this!

If I understand right, it makes polymorphic model treat `cast` and `deserialize` differenlty, where it didn't before -- specifically with regards to store keys?  Is it that `deserialize` will convert store keys but `cast` does not?

Ah but really how it does that, is it just needs to call the proper `deserialize` or `cast` on the identified model? Whereas before it always called `cast`, whether the thing was `deserialize` or `cast`, it ended up calling `cast`?

OK, this makes sense and seems right.  So it actually had nothing to do with the SerializationCoder specifically, that was me on a goose chase!

I am wondering if this is the simplest way to do make sure the proper `cast` or `deserialize` gets passed down? But it may be! If you think it is, ok!  And tests, great!  I assume these tests did _not_ pass before the change?

But yeah, it needs some more comments:

* Can you comment-document with YARD-style `@param` doc what the `operation` key is in those methods? The only legal values are :cast or :deserialize, I guess? Can you doc that?
* Maybe also comment-doc what the point of all this is -- making sure the proper `cast` or `deserialize` gets called on an underlying model class?
* If `cast` and `deserialize` are the only valid values, what do you think about, yeah, avoiding the `public_send` for just like

        if value == :cast
           type.cast(new_hash)
        elsif value == :deserialize
            type.deserialize(new_hash)
        else
            raise # ??  or should we be checking args earlier and raising ArgumentError even earlier?

   I think this might be easier to read/less mysterious/less subject to weird errors later -- it might also be a performance improvement, but I care less about that. 


Thank you for using attr_json, for getting in touch, for doing this work, and for a nice PR!

jrochkind:
(PS: If you're interested in being a committer/maintainer on this gem, let's talk, once we get this merged!)

rdubya:
Hey @jrochkind, sorry for the delay but I finally got around to making the adjustments you requested. I'm not sure the argument error is needed since these are private methods, but I threw it in there just to be safe. The tests other than the `context "with models using store keys"` passed with the existing implementation, the new context includes the test that failed before the updates.

I'd be interested in helping out with the gem but I don't have a lot of free time to dedicate to it. Some of your proposed changes do look like potentially interesting challenges to tackle though.

jrochkind:
Please no apologies for delay, we're all doing what is sustainable for us,
attr-json is on the "do it well-considered, but don't hurry" tip.

Thank you! I am myself currently out on vacation but will be back next week
and would anticipate merging this then and doing a release.

Please feel free to get in touch with any further questions or ideas!

rdubya:
Hey @jrochkind, just wanted to get this back on your radar and see if you had any other changes you'd like to see made.

jrochkind:
Thanks for the ping, no, it just slipped my radar, will try to get it shipped presently!

jrochkind:
2.1.0 has been released including this, thanks so much for your contribution @rdubya , and your patience with review/merge!

rdubya:
Awesome, thank you!

PR #248, opened 2025-10-22
allow and test on rails 8.1

### LINKED ISSUE #247: Doesn't support Rails 8.1
Rails 8.1 is out https://guides.rubyonrails.org/8_1_release_notes.html

Current attr_json version 2.5.1 doesn't support Rails 8.1
https://github.com/jrochkind/attr_json/blob/master/attr_json.gemspec#L50

Likely needs a change to support similar to
https://github.com/jrochkind/attr_json/pull/236/files

jrochkind:
Closes #247

(once we release, anyway!)


jrochkind:
lovely

